PR #43, opened 2015-09-25
Support min confidence for golint

cep21:
Fixes https://github.com/alecthomas/gometalinter/issues/42


cep21:
Not sure about appveyor but travis-ci passes.  Also the unit tests don't work on forks because you embed ```github.com/alecthomas/gometalinter/regression_tests``` which isn't the same path on a fork.  Otherwise min_confidence works.

PR #245, opened 2017-02-25
Human-readable linter output

mr-linch:
## Why?

Default output is good because it has a 	definite structure. It can easily be filtered by standard utilities or parsed by editors that does not support JSON native. 

The default output is good because it has a definite structure. It can easily be filtered by standard utilities or parsed by editors that do not support JSON natively. 

But it is not suitable for humans. Yeah, we can customize output using --format, but we can do it better. 

Let's compare:

### Default output

`gometalinter ./.`

image

### Human-readable output

`gometalinter -b ./.`

image

### What is done?

Issue-per-line output with parts highlighting.
The current color scheme works great on light/dark backgrounds.
For colorize output I used `fatih/color` package. 
It can easily be replaced by regular constants with escape codes. 

### Ideas

 - handle `--aggregate` and `--sort` and show output like eslint

image

P. S. It's my first open-source contribution and I could do something wrong.

Wait for response. Thanks.

### DIFF
--- a/config.go
+++ b/config.go
@@ -51,6 +51,8 @@ type Config struct { // nolint: aligncheck
 	EnableGC        bool
 	Aggregate       bool
 
+	Beautify bool `json:"-"`
+
 	DeadlineJSONCrutch string `json:"Deadline"`
 }
 
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/fatih/color"
 	"github.com/google/shlex"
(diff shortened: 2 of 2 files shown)


alecthomas:
I appreciate the effort, but this can be achieved with `--format` by adding ANSI escape sequences. I understand this isn't quite as convenient as a flag, but I don't generally like to add code unless there's a fairly significant reason.

Thanks again for the effort.

alecthomas:
As an example, here's doing the same thing, but with `--format`:

screen shot 2017-02-26 at 2 45 53 pm

PR #322, opened 2017-07-17
Convert default linter config from separate maps to a single map of linterConfig

dnephin:
With this change it should be easier to see the properties of a specific linter. 

I think this change will enable us to support a full linter definition from the config format. Instead of just a list of strings in the form of `name: spec`, we could support a list of boths strings and `LinterConfig` objects. That would allow a user to fully customize a linter, including setting the partition strategy.

If we read the config during `--install` it would also let them set an `InstallFrom`.

### DIFF
--- a/linters.go
+++ b/linters.go
@@ -17,7 +17,7 @@ type LinterConfig struct {
 	Pattern           string
 	InstallFrom       string
 	PartitionStrategy partitionStrategy
-	IsSlow            bool
+	IsFast            bool
 	defaultEnabled    bool
 }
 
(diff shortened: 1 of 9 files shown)


linters.go
```diff
@@ -1,30 +1,44 @@
...
 	"gopkg.in/alecthomas/kingpin.v3-unstable"
 )
 
+type LinterConfig struct {
+	Name              string
+	Command           string
+	Pattern           string
+	InstallFrom       string
+	PartitionStrategy partitionStrategy
+	IsSlow            bool
```
bradleyfalzon: Good linters are slow, there's not a lot you can do accurately with just the AST it seems. @mvdan started some work to provide interfaces to share work here https://github.com/mvdan/lint - and dominikh's megacheck helps to reuse the go-tools collection's work.

dnephin: sounds good, will swap to `IsFast`

a common interface for linters sounds like an interesting idea

dnephin:
rebased and included the change from #325

PR #392, opened 2017-11-23
Fix gas partition stategy

tytodorov:
Executing gas with argument only with the path to directory always returns that no files are analyzed:
```
$ gas some-folder
Summary:
   Files: 0
   Lines: 0
   Nosec: 0
  Issues: 0
```
In order gas to analyze the files, every file should be given as an argument.


tytodorov:
Done. The first test will cover a file on root level and the second - a file in a sub-directory.

PR #414, opened 2017-12-20
Restore the warning about updating vendored linters

orivej:
It disappeared in b0bb897a0460b9d889dce15fc2b4f64b06e817f8 when #320 moved code around:

https://github.com/alecthomas/gometalinter/pull/320/files?diff=unified#diff-7ddfb3e035b42cd70649cc33393fe32cR163


orivej:
I have noticed that moving it to `configureEnvironmentForInstall` is of better taste. Please, review again.

PR #459, opened 2018-04-09
use the standard go vet again

### LINKED ISSUE #436: Use standard `go vet` for go version >= 1.9
Related to #375

https://github.com/golang/go/issues/22296 was fixed in go1.9 (or possibly 1.10 according to the milestone, need to verify).

 `gometalinter` can check `go version` and use the standard `go vet` instead of the forked version.

bhcleek:
Fixes #436

* Remove govet
* Refactor files to be vetted so that additional expected output doesn't
  have to be added (e.g. the `root` package wasn't imported, so
  generated an additional error).
* Update the expected output to reflect the output from go1.10's `go
  vet`.


dnephin:
Thanks for the PR!

This will fix the issue for go1.9 and go1.10, but reverts the fix of #351 on go1.8 (as we can see in the CI test output). Maybe that's ok since there are options for working around the problem. If we decide it's ok to drop this fix then the test will need to be skipped for go1.8.

It also looks like go1.9 uses the old expected strings. I think the test will need an `if runtime.Version() == "go1.9"` to set a different value for `expected`.

bhcleek:
I'll make some changes and probably some force pushes while I get the tests working. I'll comment here again when it's ready.

bhcleek:
PTAL.

Note that I disabled the shadow tests for go1.8. That seems reasonable to me given that it's been over a year since go1.8 was released and there's a workaround, but I'm willing to do more work if that's an unacceptable tradeoff for gometalinter.

regressiontests/vet_test.go
```diff
@@ -1,3 +1,5 @@
+// +build go1.10
```
dnephin: Instead of using build tags to select tests could we use `runtime.Version()` to set up different expected values?  That way it should keep working when new versions are added.

regressiontests/vet_shadow_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !go1.9,go1.10
```
dnephin: Instead of using build tags to select tests could we use `runtime.Version()` to skip the test on 1.8?

```go
if runtime.Version() == "go1.8" {
    t.Skip("go vet does not have a --shadow flag in go1.8")
}
```

regressiontests/vet_test.go
bhcleek: Done. FYI, this would have kept working even once go1.11 is out, because (from https://golang.org/pkg/go/build/#hdr-Build_Constraints)
> - "go1.1", from Go version 1.1 onward
> - "go1.2", from Go version 1.2 onward
> - "go1.3", from Go version 1.3 onward
> - "go1.4", from Go version 1.4 onward
> - "go1.5", from Go version 1.5 onward
> - "go1.6", from Go version 1.6 onward
> - "go1.7", from Go version 1.7 onward
> - "go1.8", from Go version 1.8 onward
> - "go1.9", from Go version 1.9 onward
> - "go1.10", from Go version 1.10 onward

Using `runtime.Version` has the unfortunate problem of having to do string prefix comparison, because `runtime.Version()` returns the Major.Minor.Patch....

bhcleek:
PTAL

dnephin:
Thanks for confirming this fixes the problem for you.

I'm going to merge this change. It will mean that `go vet` will suffer from #351 once again, but there are workarounds (documented in that issue, as well as using v2.0.5).

leonklingele:
@dnephin works just fine for me, even with this patch applied. If I understood #351 correctly, gometalinter either didn't vet `_test.go` files or would fail if different packages were used in the same folder (e.g. `hello` and `hello_test`).

PR #513, opened 2018-08-03
fix EnableAll config in .json doesn't work

Hawken94:


main.go
```diff
@@ -223,6 +223,10 @@ Severity override map (default is "warning"):
 	include, exclude := processConfig(config)
 
 	start := time.Now()
+	if config.DefaultScope {
+		pathsArg = &[]string{"./..."}
```
dnephin: How does this relate to `EnableAll`?

PR #557, opened 2018-11-02
resolvePaths: evaluate symlinks before recursing

rbbratta:
If the root path is a symlink, evaluate it once then
recurse.  We don't want to evaluate symlinks while walking
because that can lead to loops.

The root path might be a symlink in order for developers
to place a project under GOPATH without modifying GOPATH.

This is not ideal because it rewrites the path which can confuse IDEs
and automated error parsing, but this patch at least allows us to
traverse the directories.

Signed-off-by: Ross Brattain <ross.b.brattain@intel.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -324,6 +324,12 @@ func resolvePaths(paths, skip []string) []string {
 	for _, path := range paths {
 		if strings.HasSuffix(path, "/...") {
 			root := filepath.Dir(path)
+			// if root is a symlink evaluate it once, but don't resolve symlinks when recursing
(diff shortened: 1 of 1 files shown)


rbbratta:
The bug really is in `filepath.Walk()` it shouldn't do an `os.Lstat()` on the root, because it is fine to walk from a root symlink as long as you don't evaluate symlinks later.

I can't find a better way to fix this without re-writing `filepath.Walk()` almost completely.

rbbratta:
the other fix is better, closing this.

